PR #26, opened 2016-10-13
Api stubs

jfoutz:
Minimal implementation of the spec, return types are wrong, but all of the urls can be visited, and you get a little message about what should have happened.


saurabhnanda (commented):
Would you like me to critique the code, or is it too early?

jfoutz:
Easier to fix issues now. Please indicate what should be changed so I can correct those issues in later code.

jfoutz:
And issues in this code as well!

ServantHDBC/src/Api/Authentication.hs
```diff
@@ -0,0 +1,31 @@
...
+import Control.Monad.Trans.Except
+import Servant
+
+type Auth = String
+type AuthAPI = "sessions" :>
+                         ("new" :> Post '[JSON] Auth
+                          :<|> "refresh" :> Post '[JSON] Auth
+                          :<|> "destroy" :> Post '[JSON] Auth)
+
+authNew :: ExceptT ServantErr IO Auth
```
saurabhnanda: - Is `ExceptT` mandated by servant? Or can the handler simple throw an exception? Or use something else like `EitherT`? Just curious.
- What can `ServantErr` contain? How does it get converted to JSON?

saurabhnanda: - Is there any way for input/output headers & cookies to show up in the type signature?

ServantHDBC/src/Api/Photo.hs
```diff
@@ -0,0 +1,25 @@
...
+
+import Control.Monad.Trans.Except
+import Servant
+
+type PhotoAPI = "photos" :>
+                (Capture "path_segment_1" String :>
+                 Capture "path_segment_2" String :>
+                 Capture "geometry_or_style" String :>
+                 Capture "original_filename" String :>
+                    Get '[JSON] String)
```
saurabhnanda: what does `'[JSON]` here signify - the incoming content-type or the outgoing content-type?

ServantHDBC/src/Api/Product.hs
```diff
@@ -0,0 +1,94 @@
...
+
+module Api.Product (ProductAPI, productHandlers) where
+
+import Control.Monad.Trans.Except
+import Data.Aeson
+import GHC.Generics
+import Servant
+
+
+data Product = Product
```
saurabhnanda, line 23: Is this definition correct? This seems like the definition for `ProductFilter` and not `Product`

ServantHDBC/src/Api/Product.hs
```diff
@@ -0,0 +1,94 @@
...
+instance FromJSON Product 
+
+type ProductResult = String
+type ProductAPI = "products" :>
+                (Capture "product_id" Int :>
+                    QueryParam "fields" String :>
+                    QueryParam "photo_sizes" String :>
+                    QueryParam "varants.photo_sizes" ProductResult :>
+                        Get '[JSON] String 
+            :<|> QueryParam "ids" String :>
```
saurabhnanda: Should probably parse to `[ProductId]` in Haskell

ServantHDBC/src/Api/Product.hs
```diff
@@ -0,0 +1,94 @@
...
+                 QueryParam "q" String :>
+                 QueryParam "title" String :>
+                 QueryParam "sku" String :>
```
saurabhnanda: possibly a newtype for SKU?

ServantHDBC/src/Api/Product.hs
```diff
@@ -0,0 +1,94 @@
...
+                 QueryParam "type" String :>
```
saurabhnanda: should probably parse to `ProductType` in Haskell.

ServantHDBC/src/Api/Product.hs
```diff
@@ -0,0 +1,94 @@
...
+                 QueryParam "tags" String :>
+                 QueryParam "created_at_min" String :>
+                 QueryParam "created_at_max" String :>
+                 QueryParam "updated_at_min" String :>
+                 QueryParam "updated_at_max" String :>
```
saurabhnanda: All the timestamps should probably parse to `UTCTime` in Haskell.

ServantHDBC/src/Api/Product.hs
```diff
@@ -0,0 +1,94 @@
...
+                 QueryParam "limit" String :>
```
saurabhnanda: duplicate `limit`

ServantHDBC/src/Api/Product.hs
```diff
@@ -0,0 +1,94 @@
...
+                 QueryParam "limit" Int :>
+                 QueryParam "offset" Int :>
+                 QueryParam "orderby" String :>
```
saurabhnanda: should parse to some sort of `ProductOrdering` ADT in Haskell.

ServantHDBC/src/Api/Product.hs
```diff
@@ -0,0 +1,94 @@
...
+                 QueryParam "fields" String :>
```
saurabhnanda: We can leave this out for now and refactor the code to include this later. Will probably require extensible record or suchlike for type-safe implementation.

ServantHDBC/src/Api/Product.hs
```diff
@@ -0,0 +1,94 @@
...
+    } deriving (Generic)
+
+instance FromJSON Product 
+
+type ProductResult = String
+type ProductAPI = "products" :>
+                (Capture "product_id" Int :>
+                    QueryParam "fields" String :>
+                    QueryParam "photo_sizes" String :>
+                    QueryParam "varants.photo_sizes" ProductResult :>
```
saurabhnanda: why is `ProductResult` used here?

ServantHDBC/src/Api/Tenant.hs
```diff
@@ -0,0 +1,31 @@
...
+import Control.Monad.Trans.Except
+import Servant
+
+type Tenant = String
+type TenantAPI = "tenants" :> 
+                         ("new" :> Post '[JSON] Tenant
+                          :<|> Capture "x" Int :> Get '[JSON] Tenant
+                          :<|> Capture "x" Int :> "activate" :> Post '[JSON] Tenant)
+
+tenantNew :: ExceptT ServantErr IO Tenant
```
saurabhnanda: how does one capture the incoming JSON of a POST in a type-safe manner? Shouldn't this type signature be something like:

`tenantNew :: NewTenant -> ExceptT ServantErr IO Tenant`

ServantHDBC/src/Api/Tenant.hs
```diff
@@ -0,0 +1,31 @@
...
+
+module Api.Tenant (TenantAPI, tenantHandlers) where
+
+import Control.Monad.Trans.Except
+import Servant
+
+type Tenant = String
+type TenantAPI = "tenants" :> 
+                         ("new" :> Post '[JSON] Tenant
+                          :<|> Capture "x" Int :> Get '[JSON] Tenant
```
saurabhnanda: what exactly is being captured here? what does `x` map to?

ServantHDBC/src/Api/Photo.hs
```diff
@@ -0,0 +1,25 @@
...
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeOperators #-}
+
+module Api.Photo (PhotoAPI, photoHandlers) where
+
+import Control.Monad.Trans.Except
+import Servant
+
+type PhotoAPI = "photos" :>
+                (Capture "path_segment_1" String :>
```
saurabhnanda: How is `Capture` different from `Query`? What does `path_segment_1` map to in the incoming request?

sudhirvkumar: @saurabhnanda 

the effective url is...

/photos/:path_segment_1/:path_segment-2/:geometry_or_style/:original_filename

this is what Capture does...

Query... would mean that every parameter is optional. Capture is part of the URL and mandatory... without that URL is not going to match

if we used Query then it would be...

/photos?path_segment_1=xyz&path_segment_2=abc&geometry_or_style=abc&original_filename=xyz.xyz

ServantHDBC/src/Api/Photo.hs
sudhirvkumar: JSON is the content type of both incoming & outgoing.

ServantHDBC/src/Api/Authentication.hs
sudhirvkumar: @saurabhnanda `ServantErr` is record with 4 fields... more details here 
https://hackage.haskell.org/package/servant-server-0.9.0.1/docs/Servant-Server-Internal-ServantErr.html#t:ServantErr

Yes, you need to use `ExceptT` as that what Servant is using or expecting (https://github.com/haskell-servant/servant/blob/master/servant-server/src/Servant/Server/Internal/ServantErr.hs#L22)

You can have `Header` in the Type... `Header "something" a` cookie is also part of header so I guess it can be in the type.

ServantHDBC/src/Api/Tenant.hs
sudhirvkumar: @saurabhnanda I didn't follow the whole conversation in Gitter... isn't that supposed to be done with `ReqBody contentTypes a` ?? Because that's how we are doing it and it works too! I guess its missing here..

also we can use `Handler` Monad

`tenantNew :: NewTenant -> Handler Tenant`

as

`Handler` is alias for `ExceptT ServantErr IO`

http://haskell-servant.readthedocs.io/en/stable/tutorial/Server.html#the-handler-monad

ServantHDBC/src/Api/Tenant.hs
sudhirvkumar: @saurabhnanda I guess the URL would be (/tenants/:x)   x -> is the tenant id?

I would personally would like to have these as separate lines...

```
type TenantAPI = "tenants" :> 
                         ("new" :> Post '[JSON] Tenant
                          :<|> Capture "x" Int :> Get '[JSON] Tenant
                          :<|> Capture "x" Int :> "activate" :> Post '[JSON] Tenant)
```

I like this...

```
type TenantAPI 
  = "tenants" :> "new" :> Post '[JSON] Tenant
  :<|> "tenants" :> Capture "x" Int :> Get '[JSON] Tenant
  :<|> "tenants" :> Capture "x" Int :> "activate" :> Post '[JSON] Tenant
```

if repeated "tenants" is the problem then...

```
type TenantAPI' 
  = "new" :> Post '[JSON] Tenant
  :<|> Capture "x" Int :> Get '[JSON] Tenant
  :<|> Capture "x" Int :> "activate" :> Post '[JSON] Tenant

type TenantAPI = "tenants" :> TenantAPI'
```

I like to avoid ()'s what do you think?

ServantHDBC/src/Api/Authentication.hs
jfoutz: Yes, there are Header types available. I was putting that specific part off until i got to auth, i'd rather do it at that point.

ServantHDBC/src/Api/Photo.hs
jfoutz: yes, exactly as sudkirvkumar says.

ServantHDBC/src/Api/Product.hs
jfoutz: Nope! you're right. corrected.

ServantHDBC/src/Api/Product.hs
jfoutz: I'd used a generic return, because i didn't really know what i needed at the start. Updated everything to just strings - i'll improve the returns when i start generating plausible returns.

ServantHDBC/src/Api/Product.hs
jfoutz: I added simple datatypes. It's possible to do much more sophisticated parsing. This is simply a string.

ServantHDBC/src/Api/Product.hs
jfoutz: removed, and removed from the spec.

ServantHDBC/src/Api/Tenant.hs
jfoutz: Yes, i like the TenantAPI', updated them all to match that style. no more parenthesis.

ServantHDBC/src/Api/Product.hs
jfoutz: Also, for the list parameters, i'm using the built in, ids=1&ids=2&ids=3, that seems more restful. my first cut had ids=1,2,3. it's not to bad to do parsing of a single parameter. I like the way i did it, but if you prefer a custom format for a parameter, that's pretty straightforward

ServantHDBC/src/Api/Product.hs
jfoutz: added

ServantHDBC/src/Api/Product.hs
jfoutz: added

ServantHDBC/src/Api/Product.hs
jfoutz: added

ServantHDBC/src/Api/Product.hs
jfoutz: added

ServantHDBC/src/Api/Tenant.hs
jfoutz: added the post parameters to all of the APIs

i'll need to go change all of the ExceptT ServantErr IO m, signatures to something else. I should have probably typed that out on this pass. There will at least be a ReaderT for the database added to the stack, if we decide to do something fancy with logging, that would be invoved as well.

ServantHDBC/src/Api/Authentication.hs
```diff
@@ -10,22 +10,36 @@
...
+type AuthAPI' = 
+            "new" 
+                :> ReqBody '[JSON] Auth
+                    :> Post '[JSON] String
+       :<|> "refresh" 
+                :> ReqBody '[JSON] Auth
+                    :> Post '[JSON] String
+       :<|> "destroy" 
+                :> ReqBody '[JSON] Auth
+                    :> Post '[JSON] String
```
saurabhnanda, line 33: what's the recommend way of indenting this code?

ServantHDBC/src/Api/Authentication.hs
saurabhnanda: Okay.

ServantHDBC/src/Api/Photo.hs
saurabhnanda: so, I'm not clear what does naming the path segment really do? in this case `path_segment_1` and `path_segment_2`. Because the actual handler function anyways has to give it a variable name.

ServantHDBC/src/Api/Photo.hs
```diff
@@ -10,16 +10,24 @@
...
-                 Capture "original_filename" String :>
-                    Get '[JSON] String)
+type PhotoAPI = "photos" :> PhotoAPI'
 
-photoGet :: String -> String -> String -> String -> ExceptT ServantErr IO String
+type PhotoAPI' =
+                Capture "path_segment_1" String :>
+                Capture "path_segment_2" String :>
+                Capture "geometry_or_style" String :>
+                Capture "original_filename" String :>
```
saurabhnanda, line 23: Please refer to https://github.com/vacationlabs/haskell-webapps/pull/26#discussion_r83553426

ServantHDBC/src/Api/Product.hs
```diff
@@ -12,80 +12,113 @@ module Api.Product (ProductAPI, productHandlers) where
...
 instance FromJSON Product 
 
-type ProductResult = String
-type ProductAPI = "products" :>
-                (Capture "product_id" Int :>
-                    QueryParam "fields" String :>
+data Field = Field String
+
+instance FromHttpApiData Field where
+    parseUrlPiece x = fmap Field (parseUrlPiece x :: Either Text String)
```
saurabhnanda, line 44: is the type annotation on `parseUrlPiece x` really required?

ServantHDBC/src/Api/Product.hs
```diff
@@ -12,80 +12,113 @@ module Api.Product (ProductAPI, productHandlers) where
...
+
+data ProductType = ProductType String
+
+instance FromHttpApiData ProductType where
+    parseUrlPiece x = fmap ProductType (parseUrlPiece x :: Either Text String)
+
+data ProductOrdering = ProductOrdering Int 
+
+instance FromHttpApiData ProductOrdering where
+    parseUrlPiece x = fmap ProductOrdering (parseUrlPiece x :: Either Text Int)
```
saurabhnanda, line 65: Is there any way to reduce the boilerplate related to `FromHttpApiData`?

/cc @wz1000

ServantHDBC/src/Api/Product.hs
```diff
@@ -12,80 +12,113 @@ module Api.Product (ProductAPI, productHandlers) where
...
+
+instance FromHttpApiData Sku where
+    parseUrlPiece x = fmap Sku (parseUrlPiece x :: Either Text String)
+
+data ProductType = ProductType String
+
+instance FromHttpApiData ProductType where
+    parseUrlPiece x = fmap ProductType (parseUrlPiece x :: Either Text String)
+
+data ProductOrdering = ProductOrdering Int
```
saurabhnanda, line 62: Shouldn't this be an ADT?

ServantHDBC/src/Api/Product.hs
```diff
@@ -12,80 +12,113 @@ module Api.Product (ProductAPI, productHandlers) where
...
+
+instance FromHttpApiData ProductId where
+    parseUrlPiece x = fmap ProductId (parseUrlPiece x :: Either Text Int)
+
+data Sku = Sku String
+
+instance FromHttpApiData Sku where
+    parseUrlPiece x = fmap Sku (parseUrlPiece x :: Either Text String)
+
+data ProductType = ProductType String
```
saurabhnanda, line 57: shouldn't this be an ADT?

saurabhnanda (changes requested):

ServantHDBC/src/Api/Product.hs
BartAdv: It is - maybe you meant newtype?

saurabhnanda:
@jfoutz 
- I'm not clear what part of these API stubs are known "work in progress" pieces, vs what you feel is close to final and should be reviewed more closely.
- How would you like to proceed further? Implement the Servant handlers OR implement the domain API with HDBC?

ServantHDBC/src/Api/Product.hs
saurabhnanda: I mean something like this:

```
data OrderDirection = Asc | Desc
data ProductOrdering' = Title | Sku | AdvtPrice | Weight
data ProductOrdering = ProductOrdering ProductOrdering' OrderDirection
```

ServantHDBC/src/Api/Authentication.hs
jfoutz: I don't know. The tutorial tends to put one handler on one line (http://haskell-servant.readthedocs.io/en/stable/tutorial/Server.html) That's ok for relatively few type components,  say 5. When i got to the longer methods like productFilter, having everything on one long line is hard to deal with.

Doing it this way, the path component has the least amount of indenting, it makes a table of contents kind of structure. then next level of indention are the captured variables and finally the deepest indention is the result. I can tell at a glance what i'm expecting for url components.

Finally, i put the separators first to allow easy commenting in and out of various components, much like record syntax.

I usually try to match my formatting to whatever the rest of the code base looks like. In this case, i felt i was free to do whatever worked for me. Perhaps it's not a good style, but it worked well for me, and is internally consistent across all of the files.

If there's another format you prefer, point me at an example (a long example would be nice, something like filterProducts) and i'll go back and reformat to match that style.

ServantHDBC/src/Api/Photo.hs
jfoutz: This is probably a misunderstanding on my part. I was trying to follow this part of the spec, 
https://github.com/vacationlabs/haskell-webapps/blob/master/json-api-spec.apib#L393

My belief was, you were interested in seeing how the code would differ when using path vs using query parameters. Path components are unnamed from the API consumer's point of view, but a name is still needed in the handler.

I'm not sure what change you're looking for here.

ServantHDBC/src/Api/Product.hs
jfoutz: Well, i'd argue that it is. The data declaration alone is enough to demonstrate the types are parsing out correctly. the handler simply takes a product, but it's easy to imagine taking a `(ProductType string)`. I don't know where there is an enumeration of types, but it seems clear the data declaration could be `data ProductType = ProdA | ProdB | ProdC` and the compiler will tell me the handler needs to be updated to handle the 3 variations.

_edit_

Ah, i missed this "If `product_type=physical` then `variants.weight_in_grams` and `variants.weight_display_unit` are required. On the other hand, if `product_type=digital` then the weight related field should NOT be present. Sending weight for a digital product should be an error. And NOT sending weight for a physical product should also be an error."

I'd been avoiding including any logic, just returning strings - i think you're saying you want `data ProductType = Physical | Digital`

ServantHDBC/src/Api/Product.hs
jfoutz: ok!

jfoutz:
@saurabhnanda 
- I'm not clear what part of these API stubs are known "work in progress" pieces, vs what you feel is close to final and should be reviewed more closely.

I was shooting for an 80% solution. I wanted places for everything, so i could begin implementation on the handlers. I thought problems in the API would naturally fall out as i needed to return correct results. Return types and values are obviously wrong, we want more details than "created a tenant". I can't yet say, this is the correct return for now and all time. 

On the upside, you can check this out, run it, and curl every endpoint, and get some indication of what the server should do in every case.

Some details will change, but those changes are relatively small - string to Int, or more likely Word32, responding with full json for a product. Arguments and returns will change a bit, but the names are nailed down, at this point i think it's just fleshing out what the handlers actually have to do, and adjusting the type signatures as needed. But the overall structure is in place. There shouldn't be any more handlers. There may be some helper functions, but nothing visible to the API.
- How would you like to proceed further? Implement the Servant handlers OR implement the domain API with HDBC?

I think the next step is connecting the tenant functions to postgress. Both the handler and domain api. I need (want?) the handler to drive the domain api.

saurabhnanda:
@jfoutz how quickly can we get the following domain APIs hooked to the respective JSON handlers, to have an actual working implementation:
- createTenant
- editTenant
- activateTenant
- getTenant

ServantHDBC/src/Api/Product.hs
saurabhnanda: >  i think you're saying you want data ProductType = Physical | Digital

Yes

PR #30, opened 2016-10-15
Authentication and some basic architecture 

wz1000:
**Already implemented:**
- Authentication
- Basic app architecture
- Persistent models for Tenant and User

**Goals for Fri 21/10/16:**
- Typesafe way to handle updates/edits to entities(Tenant/Users/Products etc.)
- Domain API for Tenant:
  - createTenant
  - activateTenant
  - updateTenant
  - getTenant
- Domain API for User:
  - createUser
  - editUser
  - getUser
  - activateUser
  - deactivateUser
- Experiment with type safe ways of handling roles and permissions.


ServantPersistent/src/API.hs
```diff
@@ -0,0 +1,52 @@
...
+import Data.Proxy
+import Control.Monad.Except
+import Control.Monad.Reader
+import Servant.Server.Experimental.Auth.Cookie
+import Auth
+import Types
+import DBApi
+import Models
+
+type ProductID = Int
```
saurabhnanda, line 14: `newtype ProductID = ProductID Int`? Thoughts?

ServantPersistent/src/Types.hs
```diff
@@ -0,0 +1,124 @@
...
+{-# LANGUAGE DeriveGeneric   #-}
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE DeriveAnyClass  #-}
+{-# LANGUAGE DeriveFunctor  #-}
+{-# LANGUAGE TypeInType  #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE FunctionalDependencies #-}
+{-# LANGUAGE PolyKinds  #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE GADTs  #-}
```
saurabhnanda: If there are used across all your Haskell files, you can put them in cabal, I think.

ServantPersistent/src/Types.hs
```diff
@@ -0,0 +1,124 @@
...
+    , randomSource :: RandomSource
+    , serverKey :: ServerKey
+    , environment :: Environment
+    , dbPool :: ConnectionPool
+    }
+
+type App = (ReaderT Config (ExceptT ServantErr IO))
+
+data LoginForm = Login { username :: String
+                       , password :: String
```
saurabhnanda: should we be using `Text` instead of `String` everywhere?

ServantPersistent/src/Types.hs
```diff
@@ -0,0 +1,124 @@
...
+                       } deriving (Show, Generic, Serialize, FromJSON, ToJSON)
+
+data Session = Session { sessionUser :: String
```
saurabhnanda, line 42: not clear what `sessionUser` is used for? Why is it a `String` and not a `User`?

ServantPersistent/src/Types.hs
```diff
@@ -0,0 +1,124 @@
...
+derivePersistField "UserStatus"
+
+data TenantStatus = NewT | InactiveT | ActiveT
+                        deriving (Eq, Read, Show, Generic, Serialize, FromJSON , ToJSON)
+derivePersistField "TenantStatus"
+data TSSingleton (s :: TenantStatus) where
+    TNew :: TSSingleton NewT
+    TInactive :: TSSingleton InactiveT
+    TActive :: TSSingleton ActiveT
+instance Reify TSSingleton 'NewT where
```
saurabhnanda: we definitely need a blog post to understand what's going on here and the motivation behind this.

ServantPersistent/src/Types.hs
```diff
@@ -0,0 +1,124 @@
...
+       } deriving (Generic)
+instance FromJSON TenantIdent where
+    parseJSON = genericParseJSON (defaultOptions { fieldLabelModifier = Prelude.drop 1})
+instance ToJSON TenantIdent where
+    toEncoding = genericToEncoding (defaultOptions { fieldLabelModifier = Prelude.drop 1})
+    toJSON = genericToJSON (defaultOptions { fieldLabelModifier = Prelude.drop 1})
+makeClassy ''TenantIdent
+type TenantInput = TenantIdent
+
+data TenantBase a =
```
saurabhnanda: is it easier to use TemplateHaskell to achieve this?

```
Tenant
  name Text InputOutput
  backofficeDomain Text InputOutput
  createdAt UTCTime Output
  updatedAt UTCTime Output
  tenantId TenantID Output
```

generates the following:

```
data TenantInput = TenantInput { tenantName :: Text, tenantBackofficeDomain :: Text } deriving (FromJSON)
data TenantOutput = TenantOutput { tenantName :: Text, tenantBackofficeDomain :: Text, tenantCreatedAt :: UTCTime, tenantUpdatedAt :: UTCTime, tenantId :: TenantID } deriving (ToJSON)
$(makeFieldsFor TenantInput)
$(makeFieldsFor TenantOutput)
```

saurabhnanda (commented):
Review is not complete yet. There's a lot for me to process.

ServantPersistent/src/API.hs
wz1000: Yeah, I haven't got around to implementing the model for Product yet.

ServantPersistent/src/Types.hs
wz1000: At that point I hadn't implemented the database models and just wanted to test whether authentication worked.

ServantPersistent/src/Types.hs
wz1000: This is tied to the stuff in `fromDBTenant` and `toDBTenant`.

`reify` has type `Proxy a -> s a`. In these instances, `a :: TenantStatus` and `s ~ TSSingleton`.

`TSSingleton a` is a singleton type, which means for any given `a`, there exists only one `TSSingleton a`(which is `TNew`,`TActive`,`TInactive` for `NewT`,`ActiveT`,`InactiveT` respectively)

This means that when I pattern match on the result of `reify`, GHC can generate some very useful constraints. For example

```
case reify (Proxy @s) of
  TNew -> here we get the constraint (s ~ NewT) 
            as (reify (Proxy @s) :: TSSingleton s) == (TNew :: TSSingleton NewT)
```

In the case for `TNew` on the above pattern match, we know that `s` is `NewT`. This means GHC can unify `Tenant s` with `TenantSafe NewT ()` and we can return an expression of this type. Getting a `TNew` as a result of `reify` is a proof that `s ~ NewT`

If GHC didn't know that `s ~ NewT`, then we couldn't do this because it would demand a general `Tenant s`, which we have no way of producing.

ServantPersistent/src/Types.hs
wz1000: My reasoning behind the Tenant types is this:

TenantIdent contains the fields which are always required in anything to do with tenants.

TenantBase contains a TenantIdent, timestamps(which are required most of the time) and a parameter, whose type changes based on the kind of tenant we are dealing with.

`TenantBase TenantOutputField` is wrongly named. It should be something like `TenantBase OtherDBFields`

`TenantBase TenantOutputField`, whatever its called, is isomorphic to, and should always remain isomorphic to DBTenant, which is the model of the tenant in the DB. This is evidenced by `tbdbIso`, another example of my great talent for naming things.

As of now, `TenantOutput` is `TenantBase TenantOutputField`, but that can change if, for example, we need the user as well as the tenant details in the output, or some sensitive data is added to DBTenant.

As for `TenantSafe`, we pretend it does not exist. It stores the `tenantStatus` at the typelevel, but should never be explictly used in code. Instead, `Tenant` should be used whenever a Tenant with a particular status is demanded. 

I could think of two ways to implement type safe Tenants, the first using GADTs, necessitating three seperate constructors(hence interfaces) to largely the same data. I tried this out and it seemed to be too clunky to work with.

Instead, implementing Tenant as a closed, injective type family along with TenantSafe allows us to be completely safe while having only one, conveniently polymorphic, constructor. According to me, this is much nicer than the GADT approach.

ServantPersistent/src/Types.hs
wz1000: In general, whenever you want to prove some ambiguous type to be something specific, you do it by generating a singleton value for that type, which acts as a proof.

For more on this, consult the singletons package on Hackage and the paper it references.

https://hackage.haskell.org/package/singletons

ServantPersistent/src/Types.hs
wz1000: The biggest benefit of the type family approach is that we can have intermediate TenantSafe structures which aren't proper Tenants.

ServantPersistent/src/Models.hs
```diff
@@ -0,0 +1,72 @@
...
+import Data.Text
+import Types
+import Control.Monad.IO.Class
+import Control.Monad.Reader
+import Control.Lens
+import Data.ByteString
+import Data.Proxy
+
+share [mkPersist sqlSettings { mpsGenerateLenses = True }, mkMigrate "migrateAll"] [persistLowerCase|
+DBTenant json
```
saurabhnanda, line 25: - What does `json` here do?
- Can the Haskell record be called `DBUser` while the table is called `users`?
- Please put a TODO for figuring out how to run custom SQL (triggers and check constraints) when Persistent creates or migrates a table.

ServantPersistent/src/Models.hs
```diff
@@ -0,0 +1,72 @@
...
+    username Text
+    password ByteString
+    status UserStatus
+    createdAt UTCTime
+    updatedAt UTCTime
+    UniqueUsername username
+|]
+
+
+instance HasTimestamp DBTenant where
```
saurabhnanda, line 50: Why do you need to do this? Can't classy lenses take care of this automatically?

ServantPersistent/src/DBApi.hs
```diff
@@ -0,0 +1,116 @@
...
+import Models
+import Safe
+
+type TenantOutput = TenantBase TenantOutputField
+
+data TenantOutputField = TOF { status :: TenantStatus
+                             , ownerId :: Maybe DBUserId
+                             } deriving (Generic, ToJSON)
+
+tbdbIso :: Iso' (TenantBase TenantOutputField) DBTenant
```
saurabhnanda: Is this isomorphism only for tenants, or are you expecting to use this as a pattern for every model? Will it work for `User` who's output will never contain the password field.

ServantPersistent/src/DBApi.hs
```diff
@@ -0,0 +1,116 @@
...
+tbdbIso = iso (\(TB ti c u (TOF status id)) -> DBTenant (ti ^. name) (ti ^. backofficeDomain) id status c u) (\dbt -> TB (dbt ^. tenantIdent) (dbt ^. createdAt) (dbt ^. updatedAt) (TOF (dbt ^. dBTenantStatus)  (dbt ^. dBTenantOwnerId)))
+
+dbGetTenant :: TenantID -> App (Maybe TenantOutput)
```
saurabhnanda: This needs a lot of explanation. Why the need to construct `TenantBase` and all the related types? What other alternatives were considered? Pros and cons of each?

Why can the DB API not return a DBTenant?

ServantPersistent/src/Server.hs
```diff
@@ -0,0 +1,72 @@
...
+type TestAPI = API
+
+testAPI :: Proxy TestAPI
+testAPI = Proxy
+
+newTenant :: TenantInput -> App (Headers '[Header "location" String] TenantID)
+newTenant ti = do
+    time <- liftIO $ getCurrentTime
+    let t :: Tenant NewT
+        t = liftTB $ TB ti time time ()
```
saurabhnanda: Shouldnt the DB layer be taking care of thse house keeping columns?

saurabhnanda (changes requested):
Almost through with the review.  The biggest comment is that we need a good background for what's going on with `TenantBase` `TenantStatus` `TenantIdent` etc. What is this advanced type level hackery buying us? What are the alternatives to achieve the same goal?

ServantPersistent/src/Types.hs
saurabhnanda: > TenantBase TenantOutputField, whatever its called, is isomorphic to, and should always remain isomorphic to DBTenant

Why?

saurabhnanda: I think you're tackling three things in this discussion:
- DB type, input type, and output type for tenant
- type safe implementation of tenantStatus
- Handling housekeeping columns like createdAt and updatedAt

If the solution is not tightly linked, I suggest we build one thing at a time. 

For the first point, what are the pros and cons of the current approach vs a TH based approach?

ServantPersistent/src/DBTypes.hs
```diff
@@ -0,0 +1,63 @@
...
+instance FromJSON TenantIdent where
+    parseJSON = genericParseJSON (defaultOptions { fieldLabelModifier = Prelude.drop 1})
+instance ToJSON TenantIdent where
+    toEncoding = genericToEncoding (defaultOptions { fieldLabelModifier = Prelude.drop 1})
+    toJSON = genericToJSON (defaultOptions { fieldLabelModifier = Prelude.drop 1})
+
+instance HasName TenantIdent where
+    name = lens _name (\ti n -> ti { _name = n } )
+instance HasBackofficeDomain TenantIdent where
+    backofficeDomain = lens _backofficeDomain (\ti bd -> ti { _backofficeDomain = bd } )
```
saurabhnanda, line 65: why not use the standard lens functions?

ServantPersistent/src/Server.hs
```diff
@@ -0,0 +1,64 @@
...
+type TestAPI = API
+
+testAPI :: Proxy TestAPI
+testAPI = Proxy
+
+newTenant :: TenantInput -> App (Headers '[Header "location" String] TenantID)
+newTenant ti = do
+    result <- dbCreateTenant ti
+    case result of
+         Nothing -> throwError $ err400 { errBody = "Tenant already exists" }
```
saurabhnanda, line 33: assuming that this is just a placeholder error, because this is not the only failure condition.

ServantPersistent/src/Server.hs
```diff
@@ -0,0 +1,64 @@
...
+         Just id -> return $ addHeader (show id) id
```
saurabhnanda, line 34: have you defined a custom `Show` instance for `TenantID`? How does `addHeader` really work here?

ServantPersistent/src/Domain/Tenant.hs
```diff
@@ -0,0 +1,44 @@
...
+                       , _dBTenantCreatedAt = time
+                       , _dBTenantUpdatedAt = time
+                       }
+    insertUnique dbt
+
+
+dbGetTenant :: TenantID -> App (Maybe Tenant)
+dbGetTenant = runDb . get
+
+dbGetTenantByName :: Text -> App (Maybe TenantID)
```
saurabhnanda: would it be better if this returns a `Maybe Tenant` instead of a `Maybe TenantID`?

ServantPersistent/src/Domain/Tenant.hs
```diff
@@ -0,0 +1,44 @@
...
+import Database.Persist
+import Control.Monad.IO.Class
+import Data.Text (Text)
+import Models
+import Types
+import Updater
+import DBTypes
+import Operation
+
+dbCreateTenant :: TenantInput -> App (Maybe TenantID)
```
saurabhnanda, line 17: I'm assuming the DB action monad will be handled in a future sprint.

ServantPersistent/src/DBTypes.hs
```diff
@@ -0,0 +1,63 @@
...
+data UserCreationError = UserExists Text
+                       | TenantDoesn'tExist Text
+                        deriving (Eq, Show)
+
+data Role = SiteAdmin
+          | TenantAdmin TenantID
+          | NormalUser
+
+data TenantIdent =
+    TI { _name :: Text
```
saurabhnanda: any reason why the constructor is not called `TenantIdent`?

ServantPersistent/src/DBTypes.hs
```diff
@@ -0,0 +1,63 @@
...
+
+instance HasName TenantIdent where
+    name = lens _name (\ti n -> ti { _name = n } )
+instance HasBackofficeDomain TenantIdent where
+    backofficeDomain = lens _backofficeDomain (\ti bd -> ti { _backofficeDomain = bd } )
+
+type TenantInput = TenantIdent
+
+data UserInput =
+    UserI { _firstName :: Text
```
saurabhnanda: would it be possible to follow some naming convention here? `TenantIdent` vs `UserI`?

ServantPersistent/src/DBTypes.hs
```diff
@@ -0,0 +1,63 @@
...
+          }
+
+data User =
+    User { userFirstName :: Text
+         , userLastName :: Text
+         , userEmail :: Text
+         , userPhone :: Text
+         , userUsername :: Text
+         , userTenantId :: TenantID
+         , userRole :: Role
```
saurabhnanda: this has been changed to incorporate many:many associations. A user can have many roles. A role can have many users. We can tackle this in a later sprint.

ServantPersistent/src/Domain/User.hs
```diff
@@ -0,0 +1,46 @@
...
+import Control.Monad.IO.Class
+import Control.Monad.Except
+import Models
+import Types
+import Updater
+import DBTypes
+import Domain.Tenant
+import Operation
+
+createUser :: UserInput -> App (Either UserCreationError UserID)
```
saurabhnanda: - naming consistency. `dbCreateTenant` vs `createUser`
- Also, is the pattern/interface for errors part of this sprint? `dbCreateTenant` is returning a `Maybe` whereas `createUser` is returning an `Either`, when both can have multiple reasons of failure.

ServantPersistent/src/Domain/Tenant.hs
```diff
@@ -0,0 +1,44 @@
...
+                       }
+    insertUnique dbt
+
+
+dbGetTenant :: TenantID -> App (Maybe Tenant)
+dbGetTenant = runDb . get
+
+dbGetTenantByName :: Text -> App (Maybe TenantID)
+dbGetTenantByName t = runDb $ do
+    ti <- getBy (UniqueTenant t)
```
saurabhnanda: It is not necessary that the tenant name is unique.

ServantPersistent/src/DBTypes.hs
```diff
@@ -0,0 +1,63 @@
...
+type TenantInput = TenantIdent
+
+data UserInput =
+    UserI { _firstName :: Text
+          , _lastName :: Text
+          , _email :: Text
+          , _phone :: Text
+          , _username :: Text
+          , _password :: Text
+          , _tenantName :: Text
```
saurabhnanda: why is the `tenantName` part of the `UserInput`? Shouldn't it be `TenantID` instead? Moreover, there is not guarantee that `TenantName` is unique.

ServantPersistent/src/Domain/User.hs
```diff
@@ -0,0 +1,46 @@
...
+createUser u = runExceptT $ do
+    time <- liftIO getCurrentTime
+    id <- ExceptT $ maybe (Left $ TenantDoesn'tExist (_tenantName u))
+                          Right
+                      <$> dbGetTenantByName (_tenantName u)
+    let dbu = DBUser { _dBUserFirstName = _firstName u
```
saurabhnanda: are you happy with this boilerplate?

ServantPersistent/src/DBTypes.hs
```diff
@@ -0,0 +1,63 @@
...
+type TenantOutput = DBTenant
+type UserID = Key DBUser
+
+data DBError = TenantNotFound TenantID
+             | UserNotFound UserID deriving (Eq, Show)
+data UserCreationError = UserExists Text
+                       | TenantDoesn'tExist Text
+                        deriving (Eq, Show)
+
+data Role = SiteAdmin
```
saurabhnanda: the roles are not supposed to be hard-coded. Please look at the relevant [tables in the schema](https://github.com/vacationlabs/haskell-webapps/blob/366cce31be642604077397bdcda966fa22c43686/ServantOpaleye/db/schema.sql#L59-L81) to understand that roles are defined by the tenant, as **any collection** of pre-defined permissions.

saurabhnanda (changes requested):
Stuff to be fixed in this PR:
- Style/naming related issues
- `activateTenant` is missing
- Write-up (or code comments) about the two core ideas in this PR: `Updater` and `runPermission`
- Tenant name is not supposed to be unique
- Proper usage of lenses for records (or explain why common lens function cannot be used)

Stuff to be fixed in a future sprint:
- `Role` should **not** be hard-coded to a pre-defined set of roles.
- many:many mapping between user & role
- Is there any way to force every DB action to be in the `OperationT` monad, so that no DB access is allowed without first checking the user's permissions?
- Evolve the DB monad separately from the handler monad

ServantPersistent/src/Models.hs
```diff
@@ -0,0 +1,77 @@
...
+
+instance HasHumanName DBUser where
+    firstName = dBUserFirstName
+    lastName = dBUserLastName
+instance HasContactDetails DBUser where
+    email = dBUserEmail
+    phone = dBUserPhone
+instance HasUsername DBUser where
+    username = dBUserUsername
+instance HasPassword DBUser where
```
saurabhnanda, line 70: are any of these instances used?

ServantPersistent/src/DBTypes.hs
wz1000: > there is not guarantee that TenantName is unique.

Ah. Can we change it to the backoffice domain instead of the ID? That way users won't have to lookup which ID their tenant has.

ServantPersistent/src/DBTypes.hs
wz1000: Can we have a description of the nature of roles and permissions you are looking for?

ServantPersistent/src/Domain/Tenant.hs
wz1000: I'll change that to an `Enitity Tenant` because we want the key as well.

ServantPersistent/src/Domain/Tenant.hs
wz1000: Yes, I did not realise that. I'll change that to the backoffice domain or the tenant id.

ServantPersistent/src/Domain/User.hs
wz1000: You think I should merge `TenantCreationError` and `UserCreationError`?

ServantPersistent/src/Models.hs
wz1000: `HasPassword` will be used with `Updater`, and will require a new permission for secure access. This will form the basis for reauthentication to edit protected data like passwords.

There is no direct use for `HasUsername`. I only wrote it to provide a consistent way to access all the fields of `Users`.

ServantPersistent/src/DBTypes.hs
wz1000: I'll definitely fix the naming.

ServantPersistent/src/DBTypes.hs
wz1000: Can you elaborate on what these permissions would entail?

ServantPersistent/src/DBTypes.hs
saurabhnanda: The most common use-case (the JSON API) will know the `TenantID`, right? Also, is the overall function signature better to be `createUser :: TenantID -> UserInput -> Either Err (Entity User)`?

saurabhnanda: Probably hook-up the `POST /users/new` handler to see what makes more sense?

ServantPersistent/src/Domain/User.hs
saurabhnanda: I meant the boilerplate to convert between records.

ServantPersistent/src/DBTypes.hs
saurabhnanda: Does the [roles list screen](https://vacationlabs.github.io/haskell-webapps/ui-mockups/roles.html) and [edit role screen](https://vacationlabs.github.io/haskell-webapps/ui-mockups/role-edit.html) make this clearer?

saurabhnanda:
@wz1000 are you making changes to the PR? I'd like to merge this and move to the next sprint.

Also, haven't quite understood why regular `$(makeLenses)` doesn't work in this code?

saurabhnanda:
Also, are you writing a note on `runPermission`?

ServantPersistent/src/Domain/User.hs
```diff
@@ -17,9 +17,9 @@ import Operation
 createUser :: UserInput -> App (Either UserCreationError UserID)
 createUser u = runExceptT $ do
     time <- liftIO getCurrentTime
-    id <- ExceptT $ maybe (Left $ TenantDoesn'tExist (_tenantName u))
+    id <- ExceptT $ maybe (Left $ TenantDoesn'tExist (_tenantBackofficeDomain u))
                           Right
-                      <$> dbGetTenantByName (_tenantName u)
+                      <$> dbGetTenantByBackofficeDomain (_tenantBackofficeDomain u)
```
saurabhnanda: this is fine for now. But, most probably, you will have to refactor this when you connect it to the JSON UI. The UI is anyways going to send you a tenant ID as part of the incoming JSON or as part of the URL.

ServantPersistent/src/Models.hs
```diff
@@ -29,8 +29,7 @@ DBTenant json
     status TenantStatus
     createdAt UTCTime
     updatedAt UTCTime
-    UniqueTenant name
-    UinqueBackofficeDomain backofficeDomain
+    UniqueBackofficeDomain backofficeDomain
```
saurabhnanda, line 32: did the earlier line get compiled?!

saurabhnanda (commented):
Possible to have a writeup on `runPermissions`?

ServantPersistent/src/Models.hs
wz1000: Yes

saurabhnanda:
Wow! Doesn't look too good on Persistent, I suppose!

saurabhnanda (commented):
Do you want to mix the two _type-safe update_ approaches? Or these set of commits is not about _type-safe updates_, but simple unifying the related record-types?

ServantPersistent/src/Models.hs
wz1000: There is nothing wrong with Persistent in this case. The uniqueness constraint can be called whatever you like.

wz1000:
Yes, this is about having a single representation for the related records.
Type safe updates are different and still handled exclusively by `Updater`.

ServantPersistent/src/Models.hs
wz1000: @saurabhnanda

saurabhnanda:
How does Persistent know it's a uniqueness constraint? Is that spot in the
DSL reserved only for uniqueness constraints? I though it figured it out
based on the convention of a term starting with "Unique"

saurabhnanda:
Okay. Let me know when you want me to merge this PR.

ServantPersistent/src/Models.hs
wz1000: @saurabhnanda Persistent creates constructors that represent the uniqueness constraints.

For example, if you want to get a tenant from a unique backoffice domain, you can simply use

```
getBy (UniqueBackofficeDomain "example.com")
```

instead of using something like a `select`.

wz1000:
You can merge it. I've got a few commits lined up that I don't think should
be a part of this PR.

saurabhnanda:
Do you want to implement the `undefined` functions and the fix the use of unticked promoted constructors before the merge?

wz1000:
Since we are storing the activation keys in the DB, it would require some
amount of logic. I'll prefer to deal with it in the next PR.

As for the warning for unticked promoted constructors, I would like to
disable the warning as I find code cleaner without the ticks. It's a style
issue, and there are arguments to be made either way.

See https://mail.haskell.org/pipermail/ghc-devs/2014-December/007739.html

saurabhnanda:
@wz1000 finally merged 👍 

Can you wrap this up with the following:
- What got done in this sprint
- What got pushed to the next sprint
- Any interesting learnings/roadblocks/issues/thoughts/comments (although we already have an excellent writeup from you on the `Updater` piece)

PR #35, opened 2016-10-24
Add section for UI Domain design principles

meditans:
Separating it from general design principles that should hold for all
the codebase, and from problems specific to the Database domain.

### DIFF
--- a/CONTRIBUTIONS.md
+++ b/CONTRIBUTIONS.md
@@ -63,11 +63,10 @@ A lot of people are excited about building a JSON API in Servant and everyone se
 # Higher order design goals
 
 ## General principles
-* Making nonsensical states non-representible in the domain model. eg. product can have only two type -- physical and digital. Status can have only few values. Basically a lot of ADT usage.
-* Lifting more invariants to the type-level
-* Unit tests
-* Controller tests
-* Integration tests
+* Make nonsensical states non-representible in the domain model. eg. product can have only two type -- physical and digital. Status can have only few values. Basically a lot of ADT usage.
+* Lift more invariants to the type-level.
+* Investigate how testing is (or could be) done in the various domains, re-interpreting the type of test commonly used in web development (Unit, Controller, Integration tests) if necessary, to write them in idiomatic ways.
+* Create documentation for successful workflows/techniques.
 
 ## Database domain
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTIONS.md
```diff
@@ -54,33 +54,40 @@ A lot of people are excited about building a JSON API in Servant and everyone se
...
+| Reflex FRP        | [meditans](https://github.com/meditans) |
 
 # Higher order design goals
 
+## General principles
+* Making nonsensical states non-representible in the domain model. eg. product can have only two type -- physical and digital. Status can have only few values. Basically a lot of ADT usage.
+* Lifting more invariants to the type-level
+* Unit tests
+* Controller tests
+* Integration tests
```
saurabhnanda: From a practical standpoint, right now I feel that integration tests are the **most important** aspect of UI testing. For our current project, what kind of stuff can be tested in the unit & controller tests at the **UI layer?**

CONTRIBUTIONS.md
```diff
@@ -54,33 +54,40 @@ A lot of people are excited about building a JSON API in Servant and everyone se
...
 * JSONB, ENUM, and Array support in DB library
 * Redis caching at object level
 * Redis caching at page level
-* Unit tests
-* Controller tests
-* Integration tests
+
+## UI Domain
+* Creating a reusable collection of abstractions for most used ui components (forms etc.)
+* Investigate the tradeoff of doing all the implementation in haskell vs. interface, via haskell, to html templates.
```
saurabhnanda: More stuff to think about:
- Integration with existing jQuery widgets (calendar, accordion, search/sort tables, editable grids, etc)
- Minification toolchain to reduce final JS size
- Progressive loading of JS files to reduce initial page-load time
- Server-side rendering of initial page-load
- Sharing data structures with backend
- "Adaptive validation" - it should be possible to quickly throw together an _acceptable_ form, which depends on the backend for data validation. The backend should send error messages in a structured format so that the form is able to _reasonably_ place the error messages next to the field (or fieldset) with the bad data. If the form's UX is very important, then validations can be implemented on the JS side as well, but the idea is that it **should be possible** to have a working form without client-side validation.

saurabhnanda (commented):
some more stuff to think about

PR #40, opened 2016-10-25
HDBC DomainAPI

jfoutz:
I'm interested in DomainAPI and Types

I think maybe each primitive type in Tenant and User should become a type alias, and the "create" versions should take typed arguments rather than a tennant or a user. I don't think ignoring arguments is _bad_ but it's not great.


saurabhnanda:
@jfoutz unable to compile your code because of the missing cabal file. Did you forget to commit?

jfoutz:
Yes. Added and pushed. This should cover the DomainAPI only, i'll have another update for design goals tomorrow. (at least some of the design goals will move to the next sprint.)

ServantHDBC/src/Util.hs
```diff
@@ -0,0 +1,19 @@
...
+import Database.HDBC 
+
+mkData1 f (x1:[]) = Just $ f (fromSql x1)
+mkData1 _ _ = Nothing
+mkData2 f (x1:x2:[]) = Just $ f (fromSql x1) (fromSql x2)
+mkData2 _ _ = Nothing
+mkData6 f (x1:x2:x3:x4:x5:x6:[]) = Just $ f (fromSql x1) (fromSql x2) (fromSql x3) (fromSql x4) (fromSql x5) (fromSql x6)
+mkData9 f (x1:x2:x3:x4:x5:x6:x7:x8:x9:[]) = Just $ f (fromSql x1) (fromSql x2) (fromSql x3) (fromSql x4) (fromSql x5) (fromSql x6) (fromSql x7) (fromSql x8) (fromSql x9)
+mkData9 _ _ = Nothing
+mkData11 f (x1:x2:x3:x4:x5:x6:x7:x8:x9:x10:x11:[]) = Just $ f (fromSql x1) (fromSql x2) (fromSql x3) (fromSql x4) (fromSql x5) (fromSql x6) (fromSql x7) (fromSql x8) (fromSql x9) (fromSql x10) (fromSql x11)
```
saurabhnanda, line 14: is there any way to write these functions so that they work with lists of exactly N elements? Does HDBC have functions which return fixed sized tuples instead of lists?

ServantHDBC/src/DomainApi.hs
```diff
@@ -0,0 +1,119 @@
...
+    select <- prepare conn "select id, created_at, updated_at, name, first_name, last_name, email, phone, status, owner_id, backoffice_domain from tenants where id = ? "
+    execute select [toSql id]
+    row <- fetchRow select
+    let t = fmap (mkData11 Tenant) row
+    return $ case t of
+        Just (Just t) -> Just t
+        _ -> Nothing
+
+createTenant :: Connection -> Tenant -> IO (Maybe Int)
+createTenant conn (Tenant _ created updated name first last email phone status owner backoffice)  = do
```
saurabhnanda, line 22: isn't using the record constructor a much safer way of pattern matching this? It protects against change in position of fields.

PR #42, opened 2016-10-26
Made custom types to work with insertion

sras:


SpockOpaleye/src/OpaleyeDef.hs
```diff
@@ -0,0 +1,173 @@
...
+import Data.Profunctor.Product
+import qualified Data.Profunctor.Product.Default as D
+import Data.Text
+import Data.Text.Encoding
+import Database.PostgreSQL.Simple.FromField
+import Opaleye
+
+import DataTypes
+
+type TenantTableW = (Maybe (Column PGInt4), Column PGText, Column PGText, Column PGText, Column PGText, Column PGText, Column PGText, Maybe (Column (Nullable PGInt4)), Column PGText)
```
saurabhnanda: shouldn't we be using records for this? can you quickly tell what's the difference between the first `PGText` column and the second?

SpockOpaleye/src/OpaleyeDef.hs
```diff
@@ -0,0 +1,173 @@
...
+       , required "tenant_id"
+       , required "username"
+       , required "password"
+       , optional "first_name"
+       , optional "last_name"
+       , required "status"))
+
+type RoleTableW = (Maybe (Column PGInt4), Column PGInt4, Column PGText, Column (PGArray PGText))
+
+type RoleTableR = (Column PGInt4, Column PGInt4, Column PGText, Column (PGArray PGText))
```
saurabhnanda: Create an ADT for `Permission`? Or next sprint?

SpockOpaleye/src/OpaleyeDef.hs
```diff
@@ -0,0 +1,173 @@
...
+          Just "blocked" -> UserStatusBlocked
+          _ -> error "Bad value read for user status"
+
+instance QueryRunnerColumnDefault PGText UserStatus where
+  queryRunnerColumnDefault = fieldQueryRunnerColumn
+
+instance D.Default Constant (NonEmpty Permission) (Column (PGArray PGText)) where
+  def = Constant def'
+    where
+      def' :: (NonEmpty Permission) -> (Column (PGArray PGText))
```
saurabhnanda, line 164: Interesting. `NonEmpty` list... how will you handle the case where the DB gives you an empty list of permissions?

SpockOpaleye/src/DataTypes.hs
```diff
@@ -0,0 +1,61 @@
...
+  { tenant_id :: TenantId
+  , tenant_name :: Text
+  , tenant_firstname :: Text
+  , tenant_lastname :: Text
+  , tenant_email :: Text
+  , tenant_phone :: Text
+  , tenant_status :: TenantStatus
+  , tenant_ownerid :: Maybe UserId
+  , tenant_backofficedomain :: Text
+  } deriving (Show)
```
saurabhnanda: use lenses?

SpockOpaleye/src/TenantApi.hs
```diff
@@ -0,0 +1,133 @@
...
+                                 ,tenant_status = status
+                                 ,tenant_ownerid = owner_id
+                                 ,tenant_backofficedomain = bo_domain} = do
+  ids <-
+    runInsertManyReturning
+      conn
+      tenantTable
+      (return
+         ( Nothing
+         , pgStrictText name
```
saurabhnanda: would this be easier to write with lenses?

SpockOpaleye/src/TenantApi.hs
```diff
@@ -0,0 +1,133 @@
...
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE OverloadedStrings #-}
+
+module TenantApi
+  ( create_tenant
+  , read_tenants
+  , read_tenant_by_id
+  , make_tenant
+  , remove_tenant
+  , update_tenant
```
saurabhnanda, line 13: should we be exposing the _raw_ `update_tenant` function outside this module?

SpockOpaleye/src/TenantApi.hs
```diff
@@ -0,0 +1,133 @@
...
+  return $
+    case ids of
+      [] -> Nothing
+      (x:xs) ->
+        Just $
+        tenant
+        { tenant_id = x
+        }
+
+activate_tenant :: Connection -> TenantId -> IO GHC.Int.Int64
```
saurabhnanda: need better type signatures for these functions. What's the point of returning an `Int64`?

SpockOpaleye/src/TenantApi.hs
```diff
@@ -0,0 +1,133 @@
...
+    conn
+    tenantTable
+    update_func
+    (\(id, _, _, _, _, _, _, _, _) -> id .== (constant t_tenantid))
+
+remove_tenant :: Connection -> Tenant -> IO GHC.Int.Int64
+remove_tenant conn Tenant {tenant_id = tid} = do
+  users_for_tenant <- read_users_for_tenant conn tid
+  roles_for_tenant <- read_roles_for_tenant conn tid
+  mapM_ (remove_role conn) roles_for_tenant
```
saurabhnanda, line 99: wouldn't these result in the classic N+1 bug?

SpockOpaleye/src/TenantApi.hs
```diff
@@ -0,0 +1,133 @@
...
+
+read_tenant_by_id :: Connection -> TenantId -> IO (Maybe Tenant)
+read_tenant_by_id conn id = do
+  r <- runQuery conn $ (tenant_query_by_id id)
+  return $
+    case r of
+      [] -> Nothing
+      rows -> Just $ Prelude.head $ fmap make_tenant rows
+
+make_tenant :: (TenantId, Text, Text, Text, Text, Text, TenantStatus, Maybe UserId, Text)
```
saurabhnanda: should we expose this function to the outside world? Allow anyone to make a tenant with any status?

saurabhnanda (commented):
not completed the review... just a bunch of comments.

SpockOpaleye/src/TenantApi.hs
```diff
@@ -0,0 +1,133 @@
...
+  mapM_ (remove_user conn) users_for_tenant
+  runDelete
+    conn
+    tenantTable
+    (\(id, _, _, _, _, _, _, _, _) -> id .== (constant tid))
+
+read_tenants :: Connection -> IO [Tenant]
+read_tenants conn = do
+  r <- runQuery conn $ tenant_query
+  return $ fmap make_tenant r
```
saurabhnanda: please check if you really need to do this. I also got stuck in this point, but IIUC Opaleye can convert a DB row into the corresponding Haskell data type automatically for you.

SpockOpaleye/src/UserApi.hs
```diff
@@ -0,0 +1,170 @@
...
+                           ,user_status = status} = do
+  ids <-
+    runInsertManyReturning
+      conn
+      userTable
+      (return
+         ( Nothing
+         , constant tenant_id
+         , pgStrictText username
+         , pgStrictText password
```
saurabhnanda: password as plain text?!

SpockOpaleye/src/OpaleyeDef.hs
sras: [Have changed this](https://github.com/vacationlabs/haskell-webapps/pull/42/commits/eec3307680e241f3c28553ad7f28d88321ea0535).

SpockOpaleye/src/OpaleyeDef.hs
sras: Isn't there a database constraint for that? Can't we count on it?

SpockOpaleye/src/UserApi.hs
sras: Yes. I am aware of this and have lined it up in things to take care before or in the next sprint.

SpockOpaleye/src/TenantApi.hs
sras: Yes.  Should we consider using cascading deletes in the database for this? I guess we also need to wrap these kind of logic in transactions.

SpockOpaleye/src/OpaleyeDef.hs
saurabhnanda: Yes we can. But Opaleye doesn't know that the constraint exists, does it? Is there any place in the code that throws a runtime error upon receiving an empty permission list? (This error should not occur as long as the constraint holds)

SpockOpaleye/src/TenantApi.hs
saurabhnanda: Even if we have to do this in Haskell, it should be a `DELETE...WHERE id IN ()` query

saurabhnanda:
Overall comments for this sprint:
- Any reason we aren't using records for PG types at https://github.com/vacationlabs/haskell-webapps/pull/42/files#diff-9f31a0cf6ce10f4c28c80a5ab2ff8440R17
- The indentation style is making the code hard to read (too many line breaks). Can you try what http://chrisdone.com/posts/hindent-5 does?
- Can you check if Opaleye has an implementation for `INSERT INTO... RETURNING *`? There should be a very easy way to get the inserted row back without firing another `SELECT` query. This is important for the `createdAt` and `updatedAt` columns. After every insert or update, the domain API function should return the newly-inserted or updated object back with the timestamps coming directly from the DB.

Comments for future sprints: 
- Current code is way too verbose to build anything on top of. Examples: [create_tenant](https://github.com/vacationlabs/haskell-webapps/pull/42/files#diff-49422b0d752606376d59cca995081e3bR29), [update_tenant](https://github.com/vacationlabs/haskell-webapps/pull/42/files#diff-49422b0d752606376d59cca995081e3bR79) We should probably write some TH/generic functions to solve this in the very next sprint. Or evaluate an opaleye wrapper like [dbrecord-opaleye](https://github.com/byteally/dbrecord-opaleye)
- We might need to write our own generic `FromJSON` implementation to handle cases where we want to ensure that keys with unit types `()` are absolutely omitted from the incoming JSON.
- Hook-up a digestive-functor library for JSON validation

Issue #37, opened 2016-10-24
reflex, servant-reflex - Initial sprint


meditans:
## Description

The main aim of this sprint is translating a [mock login page](https://github.com/vacationlabs/haskell-webapps/tree/master/ui-mockups) in idiomatic `reflex-frp` code.

In solving this problems three macro areas will be researched, to construct a solid foundation for future work in this area:
### How to structure an app with interoperating client and backend

To simulate the interaction between the UI and the backend, I will write a mock servant API and server. Common solutions for project structure will be examined, and I'll document the best one with a simple project.
### The validation problem

A special focus will be put in investigating the validation problem, both in Client and Server side. An optimal way to serve both needs will be researched.
### The potentiality of the frp paradigm

The frp paradigm will be followed to its natural conclusion, using the `servant-reflex` library. I'll learn to use the library, and if it's acceptabtable, use it to write transparent code in the frontend. If not, the problems in this library will be documented and reported upstream.
## Tech Stack

`reflex`, `servant`, `servant-reflex`

saurabhnanda:
Quick overview of what `servant-reflex` solves?

meditans:
Sure, I'll leave the [link to the github repo](https://github.com/imalsogreg/servant-reflex), a quick description from the readme and some of my notes:

### The problem servant-reflex solves

Keeping your frontend in sync with your API server can be difficult - when the API changes its input parameters or return type, XHR requests from the frontend will fail at runtime. If your API is defined by servant combinators, you can use servant-reflex to share the API between the server and frontend. Syncronization between is checked at compile time, and rather than building XHR requests by hand, API endpoints are available behind reflex's FRP semantics.

### Additional notes

As far as I can see this library let you abstract from the specific calls to the API: the most importan function in this library is `client`, which takes your API and returns a series of functions, one for each end point, which manage serialization, XhrRequest generation, and deserialization. And the result is presented as a `Dynamic` behaviour:

A simple example adapted from the readme: the API

``` haskell
type API = "getint"  :> Get '[JSON] Int
```

produces a function of type:

``` haskell
   getint :: MonadWidget t m
          => Event t ()                  -- ^ Trigger the XHR Request
          -> m (Event t (ReqResult Int)) -- ^ Consume the answer
```

as you can see, `getint` is a function that, every time the event passed as a first argument fires, serializes the eventual input value (here there isn't one), makes a call to the server, deserialize the response, and presents the result of the API call as an `Event`.

[...]

PR #43, opened 2016-10-28
Add mockLoginPage implementation

meditans:
Related sprint #37


UI/ReflexFRP/mockLoginPage/mockClient/Main.hs
```diff
@@ -0,0 +1,104 @@
...
+main = mainWidgetWithHead htmlHead body
+
+htmlHead :: forall t m. DomBuilder t m => m ()
+htmlHead = do
+  elAttr "meta" ("charset" =: "utf-8") (pure ())
+  elAttr "meta" ("name" =: "viewport" <> "content" =: "width=device-width, initial-scale=1.0") (pure ())
+  el "title" (text "ui-mockup")
+  styleSheet "https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css"
+  styleSheet "http://code.ionicframework.com/ionicons/2.0.1/css/ionicons.min.css"
+  styleSheet "http://localhost:8081/assets/css/Login-Form-Clean.css"
```
saurabhnanda: should be using a relative URL instead.

UI/ReflexFRP/mockLoginPage/mockClient/Main.hs
```diff
@@ -0,0 +1,104 @@
...
+  where
+    styleSheet addr = elAttr "link" ("rel"  =: "stylesheet" <> "href" =: addr) (return ())
```
saurabhnanda, line 29: don't we have a library function for this already?

meditans: It doesn't seem so. Ryan told me that I should push the things that I deem useful (like the function below in the code). This probably would also be a worthy inclusion.

BartAdv: That's tiny, but there's also 'blank' in Reflex.Dom which can be used instead of return (). Less parens, less noise ;)

UI/ReflexFRP/mockLoginPage/mockClient/Main.hs
meditans: Right! Leaving [this documentation](http://stackoverflow.com/questions/2005079/absolute-vs-relative-urls) here to explain the problem

PR #51, opened 2016-11-18
DB library comparison document

saurabhnanda:
Goal is to document our learnings after using 3 popular DB libraries in Haskell:

* Persistent
* HRR
* Opaleye


saurabhnanda:
/cc @meditans @wz1000 @sras 

Please give your feedback on the content outline. Anything else that we should be covering?

PR #52, opened 2016-11-19
Added some other design goals/questions

meditans:
Some other design goals influenced by the reflex experience

### DIFF
--- a/CONTRIBUTIONS.md
+++ b/CONTRIBUTIONS.md
@@ -104,6 +104,7 @@ A lot of people are excited about building a JSON API in Servant and everyone se
   - Minification toolchain to reduce the final JS size (closure compiler, specific ghcjs compilation options etc.)
   - Progressive loading of JS files to reduce initial page-load time
   - Server-side rendering of initial page-load
+  - Benchmarking how well the generated app fares on mobile
 
 **Please raise a PR against this file to add more higher-order design goals**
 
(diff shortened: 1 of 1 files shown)


meditans:
Also, where should I push the answer to these questions for the reflex case?

CONTRIBUTIONS.md
```diff
@@ -97,6 +97,9 @@ A lot of people are excited about building a JSON API in Servant and everyone se
   - Investigate the tradeoff of doing all the implementation in haskell vs. interface, via haskell, to html templates.
   - Investigate how to integrate with existing jQuery widgets (calendar, accordion, search/sort tables, editable grids, etc)
   - Forms relying only on server side validation or with mixed client/server side validation
+  - Investigate the preferred way to architecture an application. How powerful, and how general, is it?
```
saurabhnanda, line 100: How are you planning to answer this for reflex?

CONTRIBUTIONS.md
```diff
@@ -97,6 +97,9 @@ A lot of people are excited about building a JSON API in Servant and everyone se
   - Investigate the tradeoff of doing all the implementation in haskell vs. interface, via haskell, to html templates.
   - Investigate how to integrate with existing jQuery widgets (calendar, accordion, search/sort tables, editable grids, etc)
```
saurabhnanda, line 98: Btw, we forgot this design goal. Come back to this later?

saurabhnanda:
> Also, where should I push the answer to these questions for the reflex case?

Do something similar to #51 (markdown file in `doc` directory).

CONTRIBUTIONS.md
meditans: Yes, I plan to describe all the work I did in this area (the App datatype, with one constructor for each page of the app). That also might be turned into a tiny library, but I have to get an efficiency imprimatur from Ryan first.

CONTRIBUTIONS.md
meditans: I feel that the question of implementing all the markup directly in reflex vs using templates has been solved: I absolutely suggest using a quasiquoter, as we did.

The jQuery widgets probem is indeed more complex: probably it requires javascript ffi, that I'm learning only now. So I definitely don't have an answer for jQuery widgets integration.

PR #54, opened 2016-11-21
Add stub of content outline for ui-tutorials

meditans:
Please if you have any suggestion add them below

### DIFF
--- a/doc/ui-tutorial-structure.md
+++ b/doc/ui-tutorial-structure.md
@@ -1,0 +1,25 @@
+# Content Outline
+
+## Reflex, a starting guide
+### First part - Getting started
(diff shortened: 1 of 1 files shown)


saurabhnanda:
@meditans I'm wondering if this should be a standalone tutorial, or a comparative tutorial. In the latter, the headings/sections are common for all libraries, and are based on common requirements/tasks.

meditans:
Yes, I wondered that too; I prepared the outline this way for two main reasons:
* There are some libraries that are essential for reflex work, like `reflex-servant` which probably won't have a strict analogue
* As a reader of the tutorial, I'll probably pick one framework to read on, based on what I read online or what I want to try. I'd be distracted from the continuous context-switch.

What do you think?

PR #56, opened 2016-12-06
Send mail via sendgrid

sras:
## Sprint 3: Find and implement solution for Transactional emails.

1. Find a library that can send mails using SMTP.
2. Find a library that can be used to create multipart html emails with referable embedded media content. It should also support attachments.
3. Find a job queue that can queue any task and let the main thread continue without blocking. Use it to queue mail sending without blocking.
4. Implement a solution that uses all the above.

Time 2 days.


saurabhnanda:
We need to ensure the following:

* Attachments are supported
* Inline referencing of attachments are supported
* Job queue is generic and not specific to only emails.

saurabhnanda (changes requested):
* Can we please hook the email sending into `createTenant` domain API?
* Are we sure there isn't any easier way to set a CID for the attachment? Isn't this something that the library should do internally and expose via an API method? Can you please ask around (in-parallel) in the relevant github repo?

sras:
@saurabhnanda

You want to put the code that send activation email in the createTenant api function? Would't that be a bit out of scope of it's responsibility? 

> Isn't this something that the library should do internally and expose via an API method?

The library exposes api methods to set headers. Isn't setting specific headers trivial enough to leave those from the api?  Also, I am not sure if a general api method is even possible in this case. Because if you have a mail prepared, how will you specify a particular content block in it?

SpockOpaleye/src/Email.hs
```diff
@@ -0,0 +1,59 @@
...
+
+sendgridMail :: Mail -> IO ()
+sendgridMail mail = sendMailWithLogin' "smtp.sendgrid.net" 587 "apikey" apikey mail
+
+setCid :: T.Text -> T.Text -> Mail -> Mail
+setCid filename cid mail@Mail {mailParts = alternatives} = mail { mailParts = (setCidInAlternative filename cid) <$> alternatives}
+  where
+  setCidInAlternative :: T.Text -> T.Text -> Alternatives -> Alternatives
+  setCidInAlternative fp cid parts = (setCidForPart filename cid) <$> parts
+  setCidForPart :: T.Text -> T.Text -> Part -> Part
```
saurabhnanda: Attaching a part with a given cid (or returning an auto-generated cid), IMO, should be part of the mail library's API.

saurabhnanda:
> You want to put the code that send activation email in the createTenant api function? Would't that be a bit out of scope of it's responsibility?

If you feel that, do the following:

* createTenant
* sendActivationEmail
* createTenantAndSendEmail

saurabhnanda:
> The library exposes api methods to set headers. Isn't setting specific headers trivial enough to leave those from the api? Also, I am not sure if a general api method is even possible in this case. Because if you have a mail prepared, how will you specify a particular content block in it?

Why is it not possible for the library API to insert an attachment with a given CID? Or insert an attachment and return an auto-generated CID, guaranteed to be unique in context of the email?

SpockOpaleye/src/Email.hs
```diff
@@ -0,0 +1,69 @@
...
+    key = "cmFuZG9tdyBlaXJqd28gZWlyandvZWlyaiB3b2VyaWpvd2Vpcmogb3F3ZWlyb3F3ZWl1aHIgb3dxZXVoaXJ3b2Vpcmggb3dldWZob2kgcmV1d2ZpcmVxdWZoaXFld3VyaG9wIHF3dWh3aQ=="
+    activation_link :: T.Text
+    activation_link =  T.concat ["http://haskellwebapps.vacationlabs.com/activateTenantKey/", key]
+    from = Address { addressName = Just "VacationLabs", addressEmail = "webapps@vacationlabs.com"}
+    to = Address { addressName = Just $ T.concat [tenant_fname, " ", tenant_lname], addressEmail = tenant_email}
+    tenant_fname = (newTenant ^. firstname )
+    tenant_lname = (newTenant ^. lastname )
+    tenant_email = (newTenant ^. email )
+    makeMail :: Address -> Address -> T.Text -> IO Mail
+    makeMail from to activation_link  = simpleMail to from subject text (LT.fromStrict html) []
```
saurabhnanda: any reason for swapping the from/to argument order?

saurabhnanda: Are we sure that the `simpleMail` API necessarily needs a LazyText?

saurabhnanda: What's the last `[]` for?

SpockOpaleye/src/Email.hs
```diff
@@ -0,0 +1,69 @@
...
+      where
+      subject = "Registration Email from abc.com"
+      text = [iTrim|
```
saurabhnanda: Please move the text part also to an external template. Btw, what does `iTrim` do and which package is it coming from?

SpockOpaleye/src/Email.hs
```diff
@@ -0,0 +1,69 @@
...
+  setCidForPart :: T.Text -> T.Text -> Part -> Part
+  setCidForPart filename cid p@Part{partFilename = Just fn, partHeaders = ph} =
+    if fn == filename then p{partHeaders = (makeContentIdHeader cid):ph} else p
+  setCidForPart filename cid  p = p
+  makeContentIdHeader :: T.Text -> (ByteString, T.Text)
+  makeContentIdHeader cid = ("Content-ID", T.concat ["<", cid,  ">"])
+
+addLogo :: Mail -> IO Mail
+addLogo mail = do
+  mail_with_attachment <- addAttachment "image/png" "apple.png" mail
```
saurabhnanda: I'm not clear what's going on here. We're adding the attachment first, and after that giving it a CID? Are we sure we can't set the CID along with adding an attachment?

SpockOpaleye/src/Email.hs
```diff
@@ -0,0 +1,69 @@
...
+import           Data.String
+import           Data.String.Here
+import qualified Data.Text          as T
+import qualified Data.Text.Lazy     as LT
+import           DataTypes
+
+sendgridMail :: Mail -> IO ()
+sendgridMail mail = do
+  threadId <- forkIO $ sendMailWithLogin' "smtp.sendgrid.net" 587 "apikey" apikey mail
+  putStrLn $ show $ T.concat ["Sending main from thread ", fromString $ show threadId]
```
saurabhnanda: fix typo -- `main`

saurabhnanda (changes requested):
Overall comment:

* Please change all occurrences of `snake_case` to `camelCase`

SpockOpaleye/src/Email.hs
saurabhnanda: I looked at the PR at https://github.com/snoyberg/mime-mail/pull/48 and there indeed is a way to add an attachment along with headers (`addPart`) -- can we use that till the PR is merged and we have a top-level API to add attachment with Cid?

saurabhnanda (changes requested):

SpockOpaleye/src/Email.hs
sras: That is right. I didn't want to reimplement the attachment creation using add part because I was not sure how hard that would be to do reliably. So I used the addAttachment function and then walk through the resulting mail and look for the filename we just added, and add the content id header to that `Part`.

Now that we have seen the code of the addAttachment function, it seems quite trivial enough to just use addPart and make the attachment ourselves.

SpockOpaleye/src/Email.hs
sras: >any reason for swapping the from/to argument order?

No good reason. Just thought from -> to is a bit more natural. Didn't put too much thought into it anyway.

>Are we sure that the simpleMail API necessarily needs a LazyText?

Yes. It seems so,

https://www.stackage.org/haddock/lts-7.12/mime-mail-0.4.11/Network-Mail-Mime.html

>What's the last [] for?

It is the list of attachment specifications.

SpockOpaleye/src/Email.hs
sras: iTrim comes from the Data.String.Here module

https://hackage.haskell.org/package/here-1.2.9/docs/Data-String-Here-Interpolated.html

sras:
PR for mime-email library accepted at https://github.com/snoyberg/mime-mail/pull/48 and version updated at https://hackage.haskell.org/package/mime-mail-0.4.12/docs/Network-Mail-Mime.html

PR #60, opened 2016-12-13
Feature/error monitor

sras:


saurabhnanda:
What's the advantage of having runAppM return a AppResult? Why cant the exception handler be in runWithLogging directly? Will this remove the need for the third argument to runWithLogging as well?

PR #61, opened 2016-12-16
Feature/appcore package

sras:
I have moved some of the implementation details to a separate package called AppCore in this branch.


saurabhnanda (changes requested):
Random thoughts while looking around the codebase:

* Some helper import modules, eg. `CommonModelImport` which import and re-export the boilerplate stuff like `Database.PostgreSQL.Simple.FromField`, `Data.Profunctor.Product.Default `, `Data.Time`, `Control.Lens`, etc. Basically, stuff that needs to be imported at the top of every file. 
* Probably rename `ApiBase` to `OpaleyeHelpers` -- the former makes it sound like the core domain API.
* Is `UserDef` the actual module that will have user related APIs? If yes, then why don't we rename it to `User` itself?
* Stuff in `OpaleyeDef` is yet to be organized?
* Why does `UserId` `TenantId` stuff need to be in a different module? Also, write some TH for this boilerplate? Or perhaps talk to Tom and get this integrated into main Opaleye itself?

SpockOpaleye/src/UserApi.hs
```diff
@@ -17,49 +17,46 @@ module UserApi
...
-import           OpaleyeDef
 
-import           CryptoDef
 import           Prelude                    hiding (id)
 
-createUser :: UserIncoming -> AppM (Auditable User)
+createUser :: UserIncoming -> AppM User
 createUser user = do
   Just hash <- liftIO $ bcryptPassword $ user ^. password
   let fullUser = user { _userpolyPassword = hash }
```
saurabhnanda: use the lens setter?

SpockOpaleye/src/UserApi.hs
```diff
@@ -17,49 +17,46 @@ module UserApi
...
-  auditable <$> (createRow userTable fullUser)
+  createRow userTable fullUser
 
-updateUser :: Auditable User -> AppM (Auditable User)
+updateUser :: User -> AppM User
 updateUser user = updateAuditableRow userTable user
```
saurabhnanda, line 37: if I have an `Auditable x` and forget to call `updateAuditableRow` and instead call the regular Opaleye updat function on it, will it compile?

SpockOpaleye/src/UserApi.hs
```diff
@@ -17,49 +17,46 @@ module UserApi
...
+activateUser :: User -> AppM User
 activateUser user = setUserStatus user UserStatusActive
 
-deactivateUser :: Auditable User -> AppM (Auditable User)
+deactivateUser :: User -> AppM User
 deactivateUser user = setUserStatus user UserStatusInActive
 
-setUserStatus :: Auditable User -> UserStatus -> AppM (Auditable User)
+setUserStatus :: User -> UserStatus -> AppM User
 setUserStatus user newStatus = updateUser $ user & status .~ newStatus
```
saurabhnanda, line 46: shall we follow a convention to put an underscore at the ned of such internal APIs: `setUserStatus_`? So that if they are exported by mistake, it's easier to spot.

SpockOpaleye/src/UserApi.hs
```diff
@@ -17,49 +17,46 @@ module UserApi
...
+readUsers :: AppM [User]
+readUsers = readRow userQuery
 
-readUsersForTenant :: TenantId -> AppM [Auditable User]
-readUsersForTenant tenantId = wrapAuditable $ readRow $ userQueryByTenantid tenantId
+readUsersForTenant :: TenantId -> AppM [User]
+readUsersForTenant tenantId = readRow $ userQueryByTenantid tenantId
 
-readUserById :: UserId -> AppM (Maybe (Auditable User))
+readUserById :: UserId -> AppM (Maybe User)
```
saurabhnanda: I think this should throw an error if the user is not found. It is highly unlikely that a known UserId is missing from the DB. Think of the usage of this function -- what will the call-site really do if it gets a `Nothing` from this function?

SpockOpaleye/src/TenantApi.hs
```diff
@@ -15,38 +15,36 @@ module TenantApi
...
-import           DataTypes
 import           GHC.Int
 import           Opaleye
-import           OpaleyeDef
 import           Prelude                    hiding (id)
 import           RoleApi
 import           UserApi
 
-createTenant :: TenantIncoming -> AppM (Auditable Tenant)
+createTenant :: TenantIncoming -> AppM Tenant
```
saurabhnanda: can we add the emailing code as well?

SpockOpaleye/src/TenantApi.hs
```diff
@@ -15,38 +15,36 @@ module TenantApi
...
 createTenant tenant = do
-  auditable <$> createRow tenantTable tenant
+  createRow tenantTable tenant
 
-activateTenant :: (Auditable Tenant) -> AppM (Auditable Tenant)
+activateTenant :: Tenant -> AppM Tenant
```
saurabhnanda: should take the tenant activation key and validate it first. Or is there a different API for that?

SpockOpaleye/src/TenantApi.hs
```diff
@@ -60,16 +58,16 @@ removeTenant tenant = do
     matchFunc :: TenantTableR -> Column PGBool
     matchFunc tenant'  = (tenant' ^. id) .== (constant tid)
 
-readTenants :: AppM [Auditable Tenant]
-readTenants = wrapAuditable $ readRow tenantQuery
+readTenants :: AppM [Tenant]
```
saurabhnanda: rename to `allTenants` perhaps?

SpockOpaleye/src/RoleAPi.hs
```diff
@@ -5,39 +5,39 @@
 {-# LANGUAGE OverloadedStrings     #-}
```
saurabhnanda, line 5: check capitalization in the file name.

saurabhnanda (changes requested):

sras:
@saurabhnanda 

>Why does UserId TenantId stuff need to be in a different module?

To break cyclic imports. 

>use the lens setter?

I was not able to use a lens setter here because HasPassword class (made by TH functions) constrains the value of password to be of type `Text` and results in the error...

```
        * Couldn't match type `Data.Text.Internal.Text'
                         with `BcryptPassword'
            arising from a functional dependency between:
              constraint `HasPassword UserIncoming BcryptPassword'
```

I have now made a helper function in the AppCore that fills in the password for incoming user requests.

SpockOpaleye/src/UserApi.hs
sras: Shouldn't compile as we are unwrapping it in these functions to get the underlying type before calling opaleye's functions on it.

saurabhnanda:
>> Why does UserId TenantId stuff need to be in a different module?

> To break cyclic imports.

Please explain a little more.

> I was not able to use a lens setter here because HasPassword class (made by TH functions) constrains the value of password to be of type Text and results in the error...

Do we need a BcryptPassword <> Bytea Opaleye bridge instead?

SpockOpaleye/src/TenantApi.hs
sras: I have already made the combined process part of a UserService module...

sras:
@saurabhnanda 

> Please explain a little more.

TenantDefs need to refer to UserId data type to define ownerId field in a Tenant record. UserDefs need to refer to TenantId for defining tenantId field of the User record.

This means, if we include definitions of UserId and TenantId in UserDefs and TenantDefs, they will have to import each other. Ie UserDefs will have to import TenantDefs and TenantDefs will have to import UserDefs. This will create an import cycle and the build fails with "module imports form a cycle" error.

So by keeping the definition of TenantId and UserId in a separate module, TenantDefs and UserDefs can just import these modules, breaking the cycle.

sras:
@saurabhnanda 

>Do we need a BcryptPassword <> Bytea Opaleye bridge instead?

I am not sure. How is this related to Opaleye? The type mismatch arises from the Has* type classes generated by the makeLensesWith function.

SpockOpaleye/src/TenantApi.hs
```diff
@@ -27,58 +28,58 @@ import           Prelude                    hiding (id)
...
+  --mapM_ removeRole rolesForTenant
+  --mapM_ removeUser usersForTenant
   removeRawDbRows tenantTable matchFunc
   where
     tid = tenant ^. id
     matchFunc :: TenantTableR -> Column PGBool
     matchFunc tenant'  = (tenant' ^. id) .== (constant tid)
 
-readTenants :: AppM [Tenant]
+readTenants :: (MonadIO m, DbConnection m) =>  m [Tenant]
```
saurabhnanda, line 63: I don't think you need a `MonadIO m` constraint if you add it to the base `DbConnection m` tyepclass itself.

SpockOpaleye/appcore/src/Utils.hs
```diff
@@ -0,0 +1,10 @@
+module Utils where
+
+import UserDefs
+import CryptoDef
+import InternalUtils
+import Data.Text
+import TenantId
+
+fillPassword :: UserIncoming -> BcryptPassword -> UserPoly () () () TenantId Text BcryptPassword (Maybe Text) (Maybe Text) ()
+fillPassword incUser hash = incUser { _userpolyPassword = hash }
```
saurabhnanda, line 10: actually, I still dont understand why we can't use the lens setter for this?

saurabhnanda:
> So by keeping the definition of TenantId and UserId in a separate module, TenantDefs and UserDefs can just import these modules, breaking the cycle.

Easier to have a single package for all these ID definitions (which IMO should either be part of Opaleye itself, or should be reduced to a single TH statement by us)?

SpockOpaleye/appcore/src/Utils.hs
sras: Here is the type class that defines the `password` lens.

```
class HasPassword s a | s -> a where
  password :: Control.Lens.Type.Lens' s a
  {-# MINIMAL password #-}
        -- Defined at /vagrant/projects/haskell-webapps/SpockOpaleye/appcore/src/Lenses.hs:18:3
instance HasPassword
           (UserPoly
              key0
              created_at0
              updated_at0
              tenant_id0
              username0
              password0
              firstname0
              lastname0
              status0)
           password0
```
Doesn't it mean that the password setter can only accept a value of type password0, when used with type,

```
 (UserPoly
              key0
              created_at0
              updated_at0
              tenant_id0
              username0
              password0
              firstname0
              lastname0
              status0)

```
Doesn't the functional dependency prevents the above type from having an second instance that can accept another type for password field other than password0?

And since IncomingUser has type 'Text' for password, it follows that the setter can only accept values of type Text.

saurabhnanda: So, ensuring that only `BcryptPassword` can be set is the right thing. But that simply means that the callsite needs to convert `Text -> BcryptPassword` before using the setter. If you're making a convenience function for this -- the convenience function itself should use the setter, which is not happening in this case. And is a convenience function even required?

SpockOpaleye/src/Utils.hs
```diff
@@ -0,0 +1,9 @@
+module Utils where
+
+import AppM
+import Control.Exception
+import Exceptions
+
+returnOneIfNE :: (Monad m) => String -> [a] -> m a
+returnOneIfNE _ (x:_) = return x
+returnOneIfNE msg _ = throw $ NotFoundException msg
```
saurabhnanda, line 9: is this not compiling if replaced with https://hackage.haskell.org/package/safe-exceptions-0.1.4.0/docs/Control-Exception-Safe.html#v:throw You might have to do define a `MonadThrow` instance for `AppM`, which might not be such a bad idea as a core building block, btw.

